PR #396, opened 2022-02-23
Title1 yexxce

atuser1-scio:
Content1 yexxce


test_file_soxvtt.txt
```diff
@@ -0,0 +1 @@
+soxvtt bkpa
```
atuser1-scio, line 1: Comment1 nwubwx

PR #882, opened 2022-05-15
Title1 rmvsrh

atuser1-scio:
Content1_rmvsrh


test_file_zrtggb.txt
```diff
@@ -0,0 +1 @@
+zrtggb hufh
```
atuser1-scio, line 1: Comment1 cdncah

PR #883, opened 2022-05-15
Title2 rmvsrh

atuser1-scio:
Content2_rmvsrh


test_file_zrtggb.txt
```diff
@@ -0,0 +1 @@
+zrtggb rjiy
```
atuser1-scio, line 1: Comment2 cdncah

PR #1049, opened 2022-06-29
Title2 xkatqz

atuser1-scio:
Content2_xkatqz


test_file_srznip.txt
```diff
@@ -0,0 +1 @@
+srznip vbfs
```
atuser1-scio, line 1: Comment2 scajem

PR #1078, opened 2022-07-06
Title1 lcmged

atuser1-scio:
Content1 lcmged


test_file_figinx.txt
```diff
@@ -0,0 +1 @@
+figinx cibv
```
atuser1-scio, line 1: Comment1 zvyptv

PR #1567, opened 2023-03-17
testpr1 mrncip

atuser1-scio:
body1 mrncip


test_file_kzawni.txt
```diff
@@ -0,0 +1 @@
+kzawni dswg
```
atuser1-scio, line 1: testcomment1 aifjny

PR #1704, opened 2023-04-11
testpr1 vktmum

atuser1-scio:
body1 vktmum


test_file_hoxfrt.txt
```diff
@@ -0,0 +1 @@
+hoxfrt same
```
atuser1-scio, line 1: testcomment1 jlpilj

PR #1724, opened 2023-04-11
testpr1 fordwc

atuser1-scio:
body1 fordwc


test_file_zvtruj.txt
```diff
@@ -0,0 +1 @@
+zvtruj bzuo
```
atuser1-scio, line 1: testcomment1 nstsvy

PR #1742, opened 2023-04-14
testpr1 rcdfmo

atuser1-scio:
body1 rcdfmo


test_file_gemerx.txt
```diff
@@ -0,0 +1 @@
+gemerx ixmh
```
atuser1-scio, line 1: testcomment1 uzmatd

PR #1797, opened 2023-04-25
testpr1 bkddej

atuser1-scio:
body1 bkddej


test_file_ackkbm.txt
```diff
@@ -0,0 +1 @@
+ackkbm usfb
```
atuser1-scio, line 1: testcomment1 hofnrf

PR #1798, opened 2023-04-25
testpr1 myhyea

atuser1-scio:
body1 myhyea


test_file_afdrte.txt
```diff
@@ -0,0 +1 @@
+afdrte aldo
```
atuser1-scio, line 1: testcomment1 yctqyg

